PR #1, opened 2016-02-16
Simplify reactor-pipe RawBus inDispatcherContext ThreadLocal

CodingFabian:
The handling of "null" as initial value of the thread local makes it harder to read than necessary.
Also we can move the backpressure block into the "out of context" block, this avoids duplicate checking.


ifesdjeen:
:+1: thanks!

PR #2, opened 2016-03-01
[work-in-progress] Simplify caching registry

ifesdjeen:
Use global cache for registrations, until they get
unregistered. Performance benchmarks to follow.


ifesdjeen:
So hey @smaldini 

Could you please take a look at the current status? The commit messages are rather self-descriptive.

If we manage to merge that (or at least some similar form), pipes can go completely. I'll continue working on fluxion-bus if everything is ok here and just bring it to generally good shape.

smaldini:
Cool remember that fluxion becomes Flux. and you can get rid of dependency on reactor-stream.

PR #12, opened 2016-10-13
Scripted subscriber consumer

poutsma:
Two changes to `ScriptedSubscriber`:

 - Added consumer-based variants, as requested by @bclozel and @nebhale 
- Removed `<Predicate, Function>` expect variants, since they typically required code duplication


smaldini:
Thanks a bunch @poutsma @bclozel @nebhale

nebhale:
Yep!  This is perfect for our use-cases.  Back to nice readable testing.

PR #13, opened 2016-10-14
Removed InterruptedException on verify

poutsma:
This commit removes the `InterruptedException` declarations on `ScriptedSubscriber.verify`, and replaces them with `Thread.currentThread().interrupt()` calls.


sdeleuze:
👍 for me, I guess we can also update tests to remove the unneeded `throws InterruptedException`but that can be done later.

poutsma:
@sdeleuze, @bclozel: let me fix those two issues.

sdeleuze:
Yeah, thanks for this change :-)

PR #14, opened 2016-10-14
Allow to subscribe in ScriptedSubscriber#verify()

sdeleuze:
This change allows a much more concise syntax for tests:
 - No need to keep a reference on the ScriptedSubscriber
 - No need to explicitly specify the generic type
 - One less line to write (subscribe is done automatically)
 - Allow a clean chained syntax only

It is still possible to manually subscribe if needed.


bclozel:
Yes, please.
In many, many cases writing (and forgetting) `verify` is really cumbersome. In other cases (nested publishers), I'm happily using the first variant.

PR #17, opened 2016-10-14
Delay verification until completion

bclozel:
While using `ScriptedSubscriber` with publishers involving network
latency, I've found that existing tests started to fail.

It seems the current implementation of `ScriptedSubscriber` relies
on the `Publisher` under test to send the `onSubscribe` signal
right after subscription, whereas that signal can be sent after
some delay according to the reactive streams spec.

This commit delays all verifications after completion of the
publisher under test, including if the subscriber has been
subscribed.


smaldini:
With that commit we will definitively make a clear difference between ScriptedSubscriber and AssertSubscriber, the later being able to test advanced scenarios like infinite streams or operator behavior.

PR #19, opened 2016-10-17
Initialize TestScheduler support

smaldini:
Add TestScheduler (WIP on periodic tasks support)
Add global hooks including `TestScheduler.enable(boolean: all)` and `ScriptedSubscriber.enable(all)`
Add scripted TestScheduler#advanceTimeXxx via `ScriptedSubscriber.ValueBuilder`.

### DIFF
--- a/reactor-test/src/main/java/reactor/test/scheduler/TestScheduler.java
+++ b/reactor-test/src/main/java/reactor/test/scheduler/TestScheduler.java
@@ -1,0 +1,294 @@
+/*
+ * Copyright (c) 2011-2016 Pivotal Software Inc, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


poutsma:
What happens if any of the `advanceTime()` method variants is invoked without calling `enableVirtualTime` beforehand? Does that even work?

smaldini:
@poutsma that should fail with an IllegalStateException. Note that these advance by time are executed at verify time not at expectation time, so the exception is directly going to be raised to the user.

poutsma:
Wow, that merged quickly! Are we in a rush?

>@poutsma that should fail with an IllegalStateException

Yeah, that's not really user friendly, is it? In the original version of `ScriptedSubscriber`, I did make sure that it cannot enter illegal state, by having the separate value builder and all that. Couldn't we do something similar here: change the DSL somehow, so that throwing IllegalStateExceptions is not necessary?

smaldini:
We can't limit by typing the scope of advanceTime since it has to be setup as early and global as possible as it overrides `Schedulers` and therefore subsequent calls to `Schedulers.newSingle()` or `Schedulers.timer()`  will use `TestScheduler`.

The only thing we can do is checking if it has been assigned and eventually that could amount as an AssertionError instead of IllegalStateException.

poutsma:
Well, you could put the advanceTime logic in an extension of `ValueBuilder`, and make sure that that extension would be returned from a new static builder method. In that same method, you could enable virtual time.

These static enable/disable methods on ScriptedSubscriber are just to ugly for my taste, and I refuse to believe that there is no way around them.

smaldini:
Created a separate issue to rework API #25

PR #111, opened 2017-05-18
Upgrade to RxJava 2.1, fix Flowable conversions

akarnokd:
- Upgrade the RxJava dependency to 2.1.0
- Use `FlowableSubscriber` to not incur the spec overhead between the two libraries
- Use `Flowable` as target in the wrapper to avoid `instanceof` with the spec `subscribe` call
- `FlowableAsFluxConditionalSubscriber` subscribes to a `Flowable` and thus has to present itself as an RxJava 2 `ConditionalSubscriber`, not a Reactor 3 one

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,7 +45,7 @@ ext {
 
   // Libraries
   openHftChronicleVersion = '3.4.2'
-  rxJava2Version = '2.0.0'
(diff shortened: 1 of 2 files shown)


smaldini:
<3 @akarnokd <3

PR #132, opened 2017-11-20
#920 add clock-adapter for Scheduler interface

OlegDokuka:
see [reactor/#920](https://github.com/reactor/reactor-core/issues/920)

### DIFF
--- a/reactor-extra/src/main/java/reactor/scheduler/clock/SchedulerClock.java
+++ b/reactor-extra/src/main/java/reactor/scheduler/clock/SchedulerClock.java
@@ -1,0 +1,88 @@
+package reactor.scheduler.clock;
+
+import java.time.Clock;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.util.concurrent.TimeUnit;
+
+import reactor.core.scheduler.Scheduler;
+import reactor.util.annotation.NonNull;
+
+public class SchedulerClock extends Clock {
+
+	private final Scheduler scheduler;
+	private final ZoneId    zone;
+
+	private SchedulerClock(Scheduler scheduler, ZoneId zone) {
+		this.scheduler = scheduler;
+		this.zone = zone;
+	}
+
+	@Override
+	public ZoneId getZone() {
+		return zone;
+	}
+
+	@Override
+	public SchedulerClock withZone(ZoneId zone) {
+		return new SchedulerClock(scheduler, zone);
+	}
+
+	public Scheduler getScheduler() {
+		return scheduler;
+	}
+
+	public SchedulerClock withScheduler(Scheduler scheduler) {
+		return new SchedulerClock(scheduler, zone);
+	}
+
+	@Override
+	public long millis() {
+		return scheduler.now(TimeUnit.MILLISECONDS);
+	}
+
+	@Override
+	public Instant instant() {
+		return Instant.ofEpochMilli(millis());
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) {
+			return true;
+		}
+		if (o == null || getClass() != o.getClass()) {
+			return false;
+		}
+		if (!super.equals(o)) {
+			return false;
+		}
+
+		SchedulerClock that = (SchedulerClock) o;
+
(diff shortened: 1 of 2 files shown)


OlegDokuka:
@simonbasle @osi

reactor-extra/src/main/java/reactor/scheduler/clock/SchedulerClock.java
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	@Override
+	public long millis() {
+		return scheduler.now(TimeUnit.MILLISECONDS);
+	}
+
+	@Override
+	public Instant instant() {
+		return Instant.ofEpochMilli(millis());
```
osi: for future compatibility, this might want to use `Instant.ofEpochSecond(epochSecond, nanoAdjustment`. in Java 9 on Linux `Instant.now()` from the system `Clock` instances has microsecond granularity

OlegDokuka: Actually, I see no sense in using that API, because under the hood both methods work similarly 
``` Java
	public static Instant ofEpochSecond(long epochSecond, long nanoAdjustment) {
        long secs = Math.addExact(epochSecond, Math.floorDiv(nanoAdjustment, NANOS_PER_SECOND));
        int nos = (int)Math.floorMod(nanoAdjustment, NANOS_PER_SECOND);
        return create(secs, nos);
    }

    public static Instant ofEpochMilli(long epochMilli) {
        long secs = Math.floorDiv(epochMilli, 1000);
        int mos = (int)Math.floorMod(epochMilli, 1000);
        return create(secs, mos * 1000_000);
    }
```

OlegDokuka:
@simonbasle added docs, @osi maybe you may suggest something to add in the java-docs?

reactor-extra/src/main/java/reactor/scheduler/clock/SchedulerClock.java
osi: they work similarly, but `ofEpochMilli` is limited to millisecond-resolution.

osi:
The docs should indicate that the Clock is only capable of millisecond resolution. Otherwise looks good!

reactor-extra/src/main/java/reactor/scheduler/clock/SchedulerClock.java
OlegDokuka: Still do not get it... to extract millis from `Scheduler#now` will be needed the similar math as in `ofEpochMilli`

osi: ah i meant to ask the `Scheduler` for `now` in microseconds, and to use that.

simonbasle: so future proof as in "if at some point the Scheduler becomes capable of microsecond resolution"?

osi:
yes, exactly. 

-- 
peter royal - (on the go)

reactor-extra/src/main/java/reactor/scheduler/clock/SchedulerClock.java
OlegDokuka: I believe that we may postpone it to that changes (I mean change current implementation or even add it to Reactor later after incubation)

OlegDokuka:
@osi Provided better support for nanoseconds. Also documented that if Scheduler does not support nano, Instant time will be limited to milliseconds

reactor-extra/src/main/java/reactor/scheduler/clock/SchedulerClock.java
```diff
@@ -0,0 +1,182 @@
...
+import java.time.Clock;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.util.concurrent.TimeUnit;
+
+import reactor.core.scheduler.Scheduler;
+import reactor.util.annotation.NonNull;
+
+/**
+ * Clock adapter around {@link Scheduler}. That adoption gives batter integration with
```
simonbasle: typo "batter" -> "better"

PR #180, opened 2019-02-11
make reactor-extra a valid OSGI bundle

agjini:


pivotal-issuemaster:
@agjini Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-addons&pullRequestId=180)!

[Click here](https://cla.pivotal.io/sync/reactor?repositoryId=reactor/reactor-addons&pullRequestId=180) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@agjini Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-addons&pullRequestId=180)!

agjini:
I've moved the bnd config inside the gradle file.
It works as expected and also packages other sub projects as bundle.

However the gradle build displays warning for other projects since they not have exact same import (which is not a problem at all) :

```
> Task :reactor-adapter:jar
warning: Unused Import-Package instructions: [org.eclipse.*, javax.swing.*]

> Task :reactor-logback:jar
warning: Unused Import-Package instructions: [kotlin.*, org.eclipse.*, javax.swing.*]

```

simonbasle:
@agjini great, thanks for this contribution 👍 
This has an easy workaround when using BND to wrap the reactor-addons libraries, correct?

Asking because we're in the middle of releasing the latest release train and I don't feel comfortable merging this PR then immediately releasing it (it would be the only change in reactor-addons). So if it can be refined and merged a bit later (to give time to downstream projects to test nothing is wrong with snapshots) it would be better.

agjini:
(I'm not sure to understand your question)
For our needs, I can wait a little by deploying a local copy of reactor-extra (with osgi manifest) in our local nexus repository.

simonbasle:
@agjini see https://bnd.bndtools.org/chapters/390-wrapping.html for the potential workaround I was thinking about

agjini:
Yes I understand. It might be possible indeed. But I'm not sure how to get the work done using maven.
I've tried to wrap it using bnd but no packages were exported. So it was like a bundle without any public API.
I'm not sure of the way to use wrapping.

The workaround consisting in replacing the artifact in our repository is OK for me.

simonbasle:
ok 👍
re the warnings, would it make sense to configure this in subproject-dedicated blocks? did you notice that projects other than `reactor-extra` have specific OSGI bundling needs?

agjini:
I think it make sens to package it as osgi bundle too. The warning have no effect.

simonbasle:
@agjini sorry I didn't had time to go back to this PR earlier. We should be able to integrate it in 3.3.0.

I just wanted to know if the warnings can be eliminated by defining the `bnd` block (and most specifically the `Import-Package` instruction) individually for the 3 projects.

agjini:
Hello @simonbasle happy to hear that!

Yes I think we can define bnd block in gradle sub projects to avoid warnings.

I'll try to update the code today.

simonbasle:
@agjini without action on your part, I have made the changes I suggested a while ago and manually integrated your changes in the master branch.

agjini:
Thanks. I completely forgot do do it indeed. Thanks a lot !

-------- Message d'origine --------

PR #227, opened 2020-02-29
[bug] Remove eager "cache miss" invocation from Mono.lookup

dannyjiang001:
Hi, I found this issue while trying to use CacheMono for a project.  Mono.switchIfEmpty() eager evaluates  otherSupplier.get() which defeats the whole purpose.

PR #266, opened 2021-11-11
Avoid eager invocation of CacheFlux cacheMiss supplier

greg65236592:
Same eager loading behavior mentioned in #227 happens in `CacheFlux` too, if using reader to lookup data.

Although the CacheFlux will be deprecated soon, it would still be nice to have a fix for the existing usage. 🙂
Since this fix is very much alike with #227, I did not submit another issue for this PR, please let me know if it's needed.


simonbasle (approved):
thanks for the PR.
I actually prefer the pattern of `Flux.defer(otherSupplier)` 👍 

The other fix was `Mono.defer(() -> otherSupplier.get())` which is indeed redundant. Might make sense to also correct that other one?
I'll leave the PR open for a couple more days to give you a chance to decide on that, and otherwise will merge

PR #267, opened 2021-11-30
Align supplier invoking syntax in CacheMono

greg65236592:
#266 has been merged. This is a follow up PR for aligning supplier syntax.

Since I was a bit busy last week, I opened another PR.


simonbasle (approved):
thanks, missed that follow-up 👍

PR #285, opened 2022-04-08
fix BigInteger overflow bug on 3.4.x branch

Sunshow:
fix BigInteger overflow bug.


Sunshow:
#286 should use the more powerful fix.

